PR #16, opened 2026-08-02
Support OpenAI-compatible endpoints via OPENAI_BASE_URL

abliteration-ai:

- [x] Feature

**Related issue**: N/A

**What changed?**

`Detection/openai_config.py` — the single place where OpenAI clients are constructed — now reads the optional `OPENAI_BASE_URL` environment variable:

- `get_openai_client()` passes it to the SDK as `base_url` when set, so every consumer (`create_chat_completion`, `create_reasoning_completion`, the ADR baseline agent) can be pointed at any OpenAI-compatible API.
- `get_openai_config()` includes a `base_url` key when the variable is set.
- When `OPENAI_BASE_URL` is unset, behavior is byte-for-byte identical to before (default `api.openai.com` endpoint).

Also adds unit tests for both helpers (`Detection/tests/test_openai_config.py`) and a short README section documenting the option with a provider example (Abliteration.ai).

**Why?**

The triage model is configurable in `config_detector.yaml`, but the client was hardcoded to the hosted OpenAI API. Teams running OpenAI-compatible gateways or alternate providers (self-hosted vLLM, LiteLLM proxies, etc.) had no way to route detector traffic there. `OPENAI_BASE_URL` follows the OpenAI SDK's own convention, so it's the least surprising knob.

**How did you test it?**

- `uv run pytest tests/test_openai_config.py` — 8 passed (4 existing + 4 new covering default endpoint, custom base URL, and the config dict with/without `OPENAI_BASE_URL`).
- `black --check` clean on the touched test file; `openai_config.py` keeps the file's existing style.
- Verified no new flake8 findings in the touched code (the file's pre-existing warnings are untouched).

**Potential risks**

None expected: the new code path only activates when `OPENAI_BASE_URL` is set; existing deployments without it are unaffected. The Anthropic/Claude path is untouched.

### DIFF
--- a/Detection/README.md
+++ b/Detection/README.md
@@ -36,6 +36,17 @@ export OPENAI_API_KEY="your_key"      # For OpenAI triage LLM
 export HF_TOKEN="your_token"          # For Hugging Face models (optional)
 ```
 
+**OpenAI-compatible endpoints (optional):**
+
+Set `OPENAI_BASE_URL` to point the triage LLM at any OpenAI-compatible API. For example, to use [Abliteration.ai](https://docs.abliteration.ai):
+
+```bash
+export OPENAI_BASE_URL="https://api.abliteration.ai/v1"
+export OPENAI_API_KEY="$ABLIT_KEY"
+```
+
+Then set the triage model in `config_detector.yaml` to a model the endpoint serves (e.g. `abliterated-model`).
+
 **Models Used:**
 
 - Triage: `gpt-4o`
--- a/Detection/openai_config.py
+++ b/Detection/openai_config.py
@@ -30,9 +30,15 @@ def get_openai_client() -> OpenAI:
     Get a configured OpenAI client using OPENAI_API_KEY environment variable.
     All ads_platform components should use this function to get their OpenAI client.
 
+    Set the OPENAI_BASE_URL environment variable to point the client at any
+    OpenAI-compatible API endpoint (passed to the SDK as base_url).
+
     Returns:
         OpenAI: Configured client
     """
+    base_url = os.environ.get("OPENAI_BASE_URL")
+    if base_url:
+        return OpenAI(api_key=os.environ["OPENAI_API_KEY"], base_url=base_url)
     return OpenAI(api_key=os.environ["OPENAI_API_KEY"])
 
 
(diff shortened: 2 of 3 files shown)


pengyuzhang:
Thanks for the PR, and especially for the tests — the coverage added in `test_openai_config.py` is useful regardless of the outcome here.

We're going to close this one, for two reasons.

**1. The behavior already works without a code change.** The OpenAI Python SDK reads `OPENAI_BASE_URL` from the environment itself. Verified against the version pinned in this repo (openai 1.100.2):

```python
>>> os.environ["OPENAI_BASE_URL"] = "https://example.test/v1"
>>> OpenAI(api_key="sk-test").base_url     # no base_url argument passed
https://example.test/v1/
```

So setting the env var already points the triage LLM at any OpenAI-compatible endpoint on `main` today, and the added branch in `get_openai_client()` duplicates SDK behavior. The `get_openai_config()` change has no effect either — that function currently has no callers in the repo, so the new `base_url` key isn't consumed anywhere.

**2. The README section.** We'd prefer to keep third-party endpoint documentation vendor-neutral rather than build the example around one provider's URL, key name, and model. Separately, we'd rather not recommend running triage on an uncensored model: triage is the high-recall gate in front of the reasoning agent, and the paper's Table 2 numbers assume `gpt-4o`.

We'll document the SDK's existing `OPENAI_BASE_URL` support with a short neutral note instead.

Thanks again for taking the time to look at the project.

pengyuzhang:
Thanks for the contribution, but closing this one — the capability already exists.

The OpenAI SDK resolves `base_url` from `OPENAI_BASE_URL` natively when the argument is omitted (`openai/_client.py`, and we pin `openai>=1.76.0`), so `main`'s existing `OpenAI(api_key=...)` already honors the variable. Running the two `TestGetOpenaiClient` cases from this PR against unmodified `main`, both pass — the `get_openai_client()` change is a no-op. The same fallback applies to the `langchain_openai` paths in the benchmark agents, so the variable already redirects the project today.

The only test that fails on `main` covers `get_openai_config()`, which has no callers in the repo.

On the README: the motivation cites self-hosted vLLM and LiteLLM proxies, but the only documented example is a specific commercial endpoint. If we document this, it should be provider-neutral.

A small docs-only PR noting that `OPENAI_BASE_URL` is supported (with generic examples) would be welcome.

Issue #21, opened 2026-08-04
CursorParser and ClineParser have no Windows path support, silently returning zero telemetry on Windows


Rahul-s-007:
The top-level `README.md` claims ADR "captures agent intent, tool use, and execution traces across 7+ AI coding tools on macOS, Linux, and Windows." In practice, two of the Sensor's six parsers don't deliver on that for Windows:

- `CursorParser.__init__` (`Sensor/adr_sensor/parsers/cursor_parser.py`) only checks a macOS path (`Library/Application Support/Cursor/...`) and a Linux path (`.config/Cursor/...`).
- `ClineParser.__init__` (`Sensor/adr_sensor/parsers/cline_parser.py`) does the same, for Cline's task directory.

Neither has a Windows (`%APPDATA%`) branch. On a Windows machine, both nonexistent-path checks fail, the parser falls through to the (also nonexistent) Linux path, logs "no logs found," and returns zero entries — a silent blind spot for two of the most commonly Windows-installed tools this product claims to observe. `Sensor/README.md`'s own compatibility table (lines 15-20) is honest about the current "macOS, Linux" only support, so this isn't a doc inconsistency — it's a real, closeable gap.

Verified (two independent sources) real Windows path for Cursor: `%APPDATA%\Cursor\User\globalStorage\state.vscdb` (`C:\Users\<user>\AppData\Roaming\Cursor\User\globalStorage\state.vscdb`) — the standard Electron-app-data convention already mirrored by the existing macOS/Linux split. Cline runs as a VS-Code-extension hosted inside Cursor (per the existing paths targeting `.../Cursor/User/globalStorage/saoudrizwan.claude-dev/tasks`), so its Windows path follows the same confirmed root: `%APPDATA%\Cursor\User\globalStorage\saoudrizwan.claude-dev\tasks`.

Confirmed via a full-repo pass that no `%APPDATA%`/Windows handling exists anywhere else in `Sensor/adr_sensor/`. The other four parsers don't have this gap: `ClaudeParser`/`CodexParser` already use bare `Path.home()`-relative paths that resolve correctly on Windows with no OS branching needed, and `ClaudeDesktopParser`/`WarpParser` target apps with no Windows release to begin with.

I already have a fix scoped: add an `%APPDATA%`-based Windows branch to both parsers' path-detection logic (preserving macOS-first precedence), plus test coverage for the new OS-detection branch and a compatibility-table update in `Sensor/README.md`. Opening this first for visibility — PR to follow.

PR #25, opened 2026-08-05
fix: add Windows path support for CursorParser and ClineParser

krishnesh1:
**What changed?**
    Added Windows path support for `CursorParser` and `ClineParser`.
    - Imported the `os` module in both parser files.
    - Used `os.getenv("APPDATA")` with a fallback to `~/AppData/Roaming` to locate the `%APPDATA%` directory
  dynamically on Windows machines.
    - Added `windows_path` configuration strings matching the target application data stores for both Cursor and
  Cline.
    - Incorporated the `windows_path` directory check into the main parser `__init__` resolution chains.

    
    **Why?**
    Previously, `CursorParser` and `ClineParser` relied solely on hardcoded macOS and Linux configurations to
  discover log paths (`~/.config` and `~/Library/Application Support/`). This caused the tool to silently fail and
  report zero telemetry on Windows.

    
    **How did you test it?**
    Tested locally to ensure existing fallback chains correctly identify the local machine's OS path. Existing test
  suite checks for missing directories continue to pass as expected.

    
    **Potential risks**
    Minimal risk. The path checks (`exists()`) safely fail over through macOS -> Windows -> Linux. This change is
  purely additive and does not modify the telemetry parsing logic itself, so it shouldn't disrupt existing telemetry
  capabilities on non-Windows environments.

### DIFF
--- a/Sensor/adr_sensor/parsers/cline_parser.py
+++ b/Sensor/adr_sensor/parsers/cline_parser.py
@@ -6,6 +6,7 @@ Supports both macOS and Linux paths.
 """
 
 import json
+import os
 import re
 from datetime import datetime, timezone
 from pathlib import Path
@@ -20,7 +21,7 @@ class ClineParser(BaseParser):
     """Parser for Cline (Claude Dev) logs."""
 
     def __init__(self):
-        # Support macOS and Linux paths
+        # Support macOS, Linux, and Windows paths
         macos_path = (
             Path.home()
             / "Library/Application Support/Cursor/User/globalStorage/saoudrizwan.claude-dev/tasks"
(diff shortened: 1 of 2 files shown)


pengyuzhang:
Thanks for this, @krishnesh1 — and sorry it's ending as a close rather than a merge.

#30 landed as 89ebfe6, and it covers the same two parsers this PR fixes, so there's nothing left here to merge. Your work is in that commit, not just referenced by it: the `Co-authored-by: krishnesh1` trailer is on the squashed commit, so GitHub attributes it to you and it shows on your contribution graph. The `%APPDATA%`-with-fallback approach in #30 is the one you wrote here; it was extended to Warp (`%LOCALAPPDATA%`) and Claude Desktop, which this PR didn't cover.

One thing worth saying plainly, because it isn't a consolation prize: **your version got a detail right that the merged one gets wrong.** You resolved `windows_path` inside `__init__`, so `%APPDATA%` is read fresh every time a parser is constructed. #30 hoisted the candidates to class-level constants, which means they're evaluated once at import — so a process that sets `%APPDATA%` after importing `adr_sensor` still gets the profile-relative guess, which is exactly the #21 failure mode. #30's own test had to work around it with an `importlib.reload`. That's now a tracked follow-up on the merged PR. If you'd like to take it, the fix is to turn those class constants into a method resolved per-construction — the way `OpencodeParser._candidate_base_dirs()` already works. It'd be a clean, self-contained PR and you have the most context on it.

One place #30 did differ deliberately: `os.getenv("APPDATA", fallback)` returns `""` when the variable is set-but-empty, which then resolves relative to the CWD. #30 uses `os.environ.get("APPDATA") or fallback` so empty is treated as unset, with a test pinning that case.

Two small process notes for next time: linking the issue in the PR body (`Closes #21`) gets the PR and issue cross-referenced automatically — this one had no link, so it never showed up as a fix for #21. And the CLA check here is still unsigned, which would have blocked merging regardless; worth signing before your next PR so it isn't a surprise.

Closing as superseded by #30. Thanks again — the diagnosis was right and the fix shipped.

PR #30, opened 2026-08-08
feat(sensor): add opencode support and complete the platform matrix

### LINKED ISSUE #21: CursorParser and ClineParser have no Windows path support, silently returning zero telemetry on Windows

barisozbas:
# PR Summary — `feat/sensor-agent-and-platform-support`

**Scope:** `Sensor/` (+ the Sensor CI job and one line of the root README) · **Closes #21** · Windows env-root approach adopted from **#25** (@krishnesh1)

---

## TL;DR

Three gaps closed, all **purely additive**:

1. **opencode was entirely unsupported** — new parser, both storage backends.
2. **Windows was unreachable for four parsers** — Cursor, Cline, Warp and Claude Desktop only probed macOS/Linux paths. (Claude Code and Codex CLI already work on Windows via their platform-agnostic dotfile path — see §2.)
3. **Dispatch sessions were invisible** — delegated background agents live one directory deeper and were never discovered.

No source key, session-id format, or event `uuid` changes. Verified by diffing output against `main` on identical input: every pre-existing event is byte-identical.

---

## 1. New source: `opencode`

`Sensor/adr_sensor/parsers/opencode_parser.py` (new, 567 lines)

[opencode](https://github.com/sst/opencode) had no parser at all. The new one reads both on-disk backends:

| Backend | Detail |
| --- | --- |
| **SQLite** (current releases) | `opencode.db`, or `opencode-<channel>.db` on non-stable channels. Opened `file:...?mode=ro` so a running opencode process is never disturbed and no `-wal`/`-shm` side files are created. |
| **JSON file tree** (older releases) | A `storage/` directory of per-session/message/part JSON, in both the project-scoped and the legacy `session/info` layouts (including messages with embedded parts). |

- opencode uses the XDG layout on **every** platform, so the default is `~/.local/share/opencode` on Linux *and* macOS. `$XDG_DATA_HOME` and `$OPENCODE_DB` are honored; a macOS `Library/Application Support` fallback is probed defensively.
- Tool calls are classified against opencode's built-in registry. Anything else containing an underscore is recorded as `tool_type: "mcp_tool"` with `server_name` split out, matching how opencode namespaces MCP tools as `<server>_<tool>`.
- Text, reasoning, subtask and file parts become conversation content; structural parts (`step-start`, `snapshot`, `patch`, `retry`, `compaction`, …) are skipped.
- Session ids strip opencode's own `ses_` prefix, so you get `opencode_abc123` rather than `opencode_ses_abc123`.

## 2. Windows support — the platform matrix was incomplete

Closes #21. Four parsers silently returned nothing on Windows because they only probed macOS and Linux locations. Each now resolves against an ordered candidate list rather than an either/or pair:

| Parser | Windows root | Path added |
| --- | --- | --- |
| `cursor` | `%APPDATA%` | `Cursor/User/globalStorage/state.vscdb` |
| `cline` | `%APPDATA%` | `Cursor/User/globalStorage/saoudrizwan.claude-dev/tasks` |
| `warp` | `%LOCALAPPDATA%` | `warp/Warp/data/warp.sqlite` |
| `claude_desktop` | `%APPDATA%` | `Claude/local-agent-mode-sessions` |

The Windows roots are resolved from the environment first (`utils/platform_paths.py`), with the profile-relative `~/AppData/...` location only as fallback — on roaming-profile and redirected-folder setups `%APPDATA%`/`%LOCALAPPDATA%` live outside the user profile, and a hardcoded guess finds nothing. This adopts the env-first approach from #25 by @krishnesh1 (credited as commit co-author) and extends it to Warp and Claude Desktop, which #25 didn't cover. It also matches how the parsers already honor `XDG_DATA_HOME` / `XDG_CACHE_HOME` / `OPENCODE_DB`.

**Warp also gains the sandboxed macOS group-container path** — `~/Library/Group Containers/2BBY89MBSN.dev.warp/Library/Application Support/dev.warp.Warp-Stable/warp.sqlite`. This is where the database actually lives on current Warp builds, so the previous single path meant Warp produced zero events for most macOS users. This shows up in the regression diff below as a genuinely new source.

**`claude` and `codex` needed no change.** Both use a single dotfile path (`~/.claude/projects`, `~/.codex/sessions`) with no OS-specific branching — that convention already covers Windows, since `Path.home()` resolves correctly there via `USERPROFILE`/`HOMEDRIVE`+`HOMEPATH`. The platform column for both now reads `macOS, Linux, Windows`, same as the four parsers above.

## 3. Claude Desktop agent mode: Dispatch sessions

Dispatch (delegated background agents) stores sessions one level deeper, under `agent/local_ditto_<uuid>/`, and the old discovery walk never looked there. They now surface under the **existing** `claude_desktop` source:

- session-id prefix `claude_desktop_dispatch_` (interactive ids unchanged)
- `is_dispatch: true` in `session_context`

That split matters for detection: these are agents running unattended, which is exactly the population you least want a blind spot in, and they should be scored differently from interactive sessions.

`session_context` also now carries `plugins`, `skills` and `claude_code_version` from the `system:init` event, plus `session_type`, `cli_session_id`, the memory/skills/plugins toggles and the available slash commands.

> Deliberately **not** captured: the `accountName` / `emailAddress` fields present in session metadata. They add per-user PII beyond the `username`/`hostname` the Sensor already collects, and nothing downstream needs them.

## 4. Observer / CLI cleanup

`ingest_all()` now iterates a `SOURCES` table and resolves each parser as `self.<source>_parser`, replacing seven near-identical `if source_filter in [...]` blocks with one loop. Platform-only agents are declared in `PLATFORM_RESTRICTED_SOURCES` instead of inline `platform.system()` checks, and the CLI derives its `--source` choices from `SOURCES`.

This is not cosmetic: adding these sources under the old structure meant three more copies of the same eight-line block. It is also what lets `CONTRIBUTING.md` honestly describe adding a parser as a two-line registration.

## 5. Environment

- **CI:** the Sensor job runs a **Python 3.9 / 3.10 / 3.11 / 3.12 / 3.13** matrix — every version advertised in the package classifiers. The package has always declared `requires-python = ">=3.9"` but only 3.12 was ever exercised. `uv build` still runs once, on 3.12.
- **`pyproject.toml`:** added the 3.10 and 3.13 classifiers (3.10 was supported but unlisted) and OS classifiers for macOS, Linux and Windows.

## 6. Docs

Supported-agent table now carries source keys and an accurate per-agent platform column; new sections for Dispatch and opencode; a documented `session_context` payload; an Environment section covering supported runtimes and the three environment variables the Sensor reads (`XDG_CACHE_HOME`, `XDG_DATA_HOME`, `OPENCODE_DB`); and a rewritten "Adding a New Parser" guide in both README and `CONTRIBUTING.md`. `examples/demo.py` gained opencode and Dispatch sample events.

`CONTRIBUTING.md` now states explicitly that source keys are part of the Sensor's public contract and that renaming one is a breaking change to be avoided.

---

### Unit tests — 114 passed (64 pre-existing, 50 new)

No existing test needed modification. New coverage:

- **opencode:** backend detection (XDG, `$OPENCODE_DB`, channel-suffixed DBs, SQLite-over-JSON priority), SQLite parsing with tool calls / MCP classification / error states / reasoning-subtask-file parts / age filtering / truncation, both JSON layouts, tool classification.
- **Claude Desktop:** interactive and Dispatch sessions, mixed discovery, age filtering, missing and malformed metadata, malformed audit lines.
- **Platform matrix:** an explicit test per parser asserting every OS its agent ships on is present in the candidate list, plus fallback-ordering behaviour. These are the tests that would have caught the original Windows gap.
- **Windows app-data resolution:** `%APPDATA%`/`%LOCALAPPDATA%` set, unset, and empty-string cases, plus an end-to-end test that places a Cursor DB under a redirected `%APPDATA%` outside the user profile and asserts discovery — the #21 scenario.

### Real executions

**Against this machine's actual agent logs** (`~/.claude/projects`, `~/.codex/sessions`):

```
| CLAUDE   |  3 conversations | 197 messages | 147 tool calls |
| CODEX    |  3 conversations |  18 messages |   0 tool calls |
```

**Against a synthetic home with every agent at its real default path**, driven through the installed `adr-sensor` CLI (not pytest), posing as macOS:

```
| CLAUDE         | 1 | 2 | 1 |
| CLAUDE_DESKTOP | 2 | 4 | 2 |   <- interactive + Dispatch
| WARP           | 1 | 1 | 0 |   <- via the group-container path
| CODEX          | 1 | 3 | 1 |
| OPENCODE       | 1 | 2 | 2 |
```

Output payload assertions, all passing: Dispatch prefix present · interactive prefix unchanged · thinking blocks excluded · opencode MCP tool classified with `server_name` · opencode built-in not misclassified · `ses_` prefix stripped · tool results back-filled onto their calls.

**Windows-shaped home:** all four parsers resolved to their `AppData` locations and ingested successfully.

**Other CLI modes:** `--save-sessions` wrote 6 session files and correctly filtered all 6 as already-seen on a second run; `--output-format jsonl` produced 6 valid JSONL lines; opencode's legacy JSON backend, channel-suffixed DB auto-detection and `$OPENCODE_DB` override each verified through the CLI; `opencode.db` md5 unchanged after parsing, with no `-wal`/`-shm` side files.

### Regression check against `main`

Installed `main` in a separate venv and ran both versions against the same synthetic home:

```
session_ids REMOVED by this PR : none
sources     REMOVED by this PR : none
events present in both         : identical on source, model, project_path, uuid and chat_history
added                          : claude_desktop_dispatch_*, opencode_*, warp_*
```

The `uuid` is a SHA-256 content hash, so identical uuids mean downstream dedup keys are stable.

### Detection component

`Detection/` has **no coupling to the Sensor** — it never imports `adr_sensor` and never reads its output. (The `"claude"` string literals there are the Claude CLI binary name in `subprocess` calls, and `sensor_data_processor` is an unrelated IoT benchmark MCP server.) Nothing in this PR can affect it.

### Static checks

`ruff check` passes clean on `adr_sensor/`, `tests/test_parsers.py` and `examples/demo.py`. All files parse under Python 3.9 grammar (`ast.parse(..., feature_version=(3, 9))`); annotations use `typing.List`/`Dict`/`Optional` throughout.

---

## Risks

- **Low.** The regression diff shows the change is additive; no consumer-visible identifier moves.
- New event volume: sites using Claude Desktop Dispatch, opencode, or Warp on macOS will see sources appear that previously produced nothing. That is the intent, but it is a volume change worth noting for anyone with ingestion quotas.
- All parsers are read-only, and the opencode SQLite connection is explicitly read-only.
- The CI matrix grows the Sensor job from one Python version to five. If any version proves troublesome on the runners, the matrix is the first thing to trim.


pengyuzhang (changes requested):
Strong PR — the opencode parser, the Warp Group Container path, and the CI matrix are all clearly the product of testing against real installs. Verified locally: `uv run pytest tests/ -q` → **108 passed** (up from 64, 38 new tests), `uv run ruff check adr_sensor/` clean, and the changes are purely additive with existing paths retained as fallbacks.

Three things before merge — one of which is a community question rather than a code one.

---

**1. This overlaps #25, and #25 is more correct on the line they both touch.**

@krishnesh1 opened #25 on Aug 5 (addressing @Rahul-s-007's #21) with the same Cursor/Cline Windows fix. Functionally #30 fully subsumes it — plus Warp, Claude Desktop, opencode, and 38 tests that #25 doesn't have. But #25 resolves the Windows app-data root from the environment, and this PR hardcodes it:

```python
# #25
Path(os.getenv("APPDATA", Path.home() / "AppData/Roaming")) / "Cursor/User/globalStorage/state.vscdb"

# #30
Path.home() / "AppData/Roaming" / _CURSOR_STORAGE_SUFFIX
```

On roaming profiles and redirected-folder setups — common in exactly the enterprise Windows fleets this parser targets — `%APPDATA%` is not under `~/AppData/Roaming`, so the parser silently finds nothing. That's the same failure mode #21 reports. Same applies to `AppData/Local` in `WarpParser`, which should read `%LOCALAPPDATA%`.

Suggested:

```python
_APPDATA = Path(os.environ.get("APPDATA") or Path.home() / "AppData/Roaming")
_LOCALAPPDATA = Path(os.environ.get("LOCALAPPDATA") or Path.home() / "AppData/Local")
```

The parsers already read `XDG_DATA_HOME`, `XDG_CACHE_HOME`, and `OPENCODE_DB` from the environment, so this is also more consistent with the surrounding code. Please credit #25 when you adopt it — I'd like the external contribution acknowledged rather than quietly superseded.

**2. The CI matrix skips 3.10, but `pyproject.toml` now advertises it.**

This PR adds the `Programming Language :: Python :: 3.10` and `:: 3.13` classifiers; the matrix runs 3.9/3.11/3.12/3.13. With `adr-sensor` heading to PyPI, we shouldn't claim a version we don't test — either add `"3.10"` to the matrix or drop the classifier.

**3. `license/cla` is still pending** on this PR.

---

Points 2 and 3 are quick. Point 1 is the one I'd hold on: the current form is a small regression against #25 on the specific bug being fixed, and I'd rather not merge a Windows fix that misses redirected app-data roots.

barisozbas:
Thanks for the careful review — all three points addressed in 9e52e31.

**1. Env-resolved app-data roots, adopting #25.** Agreed — the profile-relative paths were a regression against #25 on redirected/roaming profiles, which is the exact failure mode #21 reports. All four Windows-touching parsers now resolve their roots from the environment first, via a shared `utils/platform_paths.py`:

```python
def windows_appdata() -> Path:
    return Path(os.environ.get("APPDATA") or Path.home() / "AppData/Roaming")

def windows_local_appdata() -> Path:
    return Path(os.environ.get("LOCALAPPDATA") or Path.home() / "AppData/Local")
```

Cursor, Cline and Claude Desktop use `%APPDATA%`; Warp uses `%LOCALAPPDATA%`. (`or` rather than a `get()` default so an empty-string env var also falls back.) This is @krishnesh1's approach from #25, credited as co-author on the commit, and extends it to the two parsers #25 didn't cover — so together with #21's report this closes the redirected-profile gap across the board. New tests cover set/unset/empty-string resolution for both variables, plus an end-to-end test that places a Cursor DB under a redirected `%APPDATA%` outside the user profile (`fileserver/profiles$/alice/...`) and asserts the parser finds it.

**2. 3.10 added to the Sensor CI matrix** — all five advertised versions (3.9–3.13) now run; the `uv build` step still runs once on 3.12.

**3. CLA** — signed; `license/cla` is green.

Suite is now 114 passing (6 new for the env resolution), `ruff check` clean.

pengyuzhang (approved):
Approving — the ingestion logic is sound and the verification work behind this is unusually thorough.

Re-verified at `bac303d`: 114 tests pass locally, `ruff check adr_sensor/ tests/test_parsers.py examples/demo.py` clean, and the whole matrix (py3.9–3.13) plus the new `test-sensor-gate` aggregate check is green. I also confirmed the central claim independently: `AgentEvent.__post_init__` hashes device/user/timestamp/source/session_id/message-count/tool-count/content-prefix and **not** `session_context`, so enriching `session_context` genuinely cannot move existing `uuid`s. The `if: always()` guard on the aggregate gate is right — skipped jobs otherwise satisfy required checks.

Merging as-is. Two items from my earlier review are still open; neither is a correctness risk to existing telemetry, so they're fine as follow-ups rather than blockers:

1. **Windows app-data roots are captured at import time.** `CursorParser.DB_PATHS`, `ClineParser.BASE_PATHS`, `WarpParser.DB_PATHS` and `DEFAULT_BASE_PATHS` evaluate `windows_appdata()` / `Path.home()` at module import, so a process that sets `%APPDATA%` after importing `adr_sensor` still gets the profile-relative guess — the #21 failure mode. `OpencodeParser._candidate_base_dirs()` already does this correctly (resolved per-construction); the four Windows parsers should match it. Worth noting #25 got this part right by keeping resolution in `__init__`. Fixing it also lets `test_redirected_appdata_end_to_end` drop the `importlib.reload` workaround.

2. **`opencode_parser.py:175` interpolates the path into a SQLite URI unencoded.** `f"file:{db_path}?mode=ro"` fails with `unable to open database file` if the path contains a `%` (verified locally); the exception is caught at line 194 and the source silently yields zero events. `"file:" + pathname2url(str(db_path)) + "?mode=ro"` fixes it. Matters more than it looks since `opencode` isn't in `PLATFORM_RESTRICTED_SOURCES` and so runs on Windows, where backslash paths go into the URI too.

Smaller follow-ups: `test_first_candidate_used_when_none_exist` asserts against the developer's real `$HOME` without isolating it (fails on a Linux box that actually has Cursor installed); `PLATFORM_RESTRICTED_SOURCES` covers only `claude_desktop` while the README promises automatic skipping more broadly, and a skipped source `continue`s silently with no explanation to the user; and `_classify_tool`'s "underscore ⇒ MCP" heuristic will attach a fabricated `server_name` to snake_case non-builtin tools, which is a precision loss in a field detection rules key on.

Also: **no Windows runner.** The headline feature is Windows support and every Windows assertion is a simulation on ubuntu. One `windows-latest` leg would exercise real `Path.home()`, path separators, and the sqlite URI above — I'd trade two Python versions for it.

Thanks for crediting @krishnesh1 as co-author. #25 is fully subsumed by this and should be closed as superseded; it also links no issue, whereas this correctly closes #21.

PR #38, opened 2026-08-10
fix(sensor): attribute MCP servers and failed tool calls in Claude Code sessions

aglanio:
## Problem

The Claude Code parser loses three signals that matter for downstream detection:

1. `tool_type` is hardcoded to `"tool_use"` for every tool call ([claude_parser.py#L279](https://github.com/uber/ADR/blob/main/Sensor/adr_sensor/parsers/claude_parser.py)), so terminal commands and MCP calls end up in the same bucket.
2. `server_name` is never populated, even though Claude Code namespaces MCP tools as `mcp__<server>__<tool>` and the server is recoverable from the name alone.
3. The `is_error` flag on `tool_result` is never read, so every call is recorded as `success` or `unknown` — failed and blocked tool calls disappear.

Measured on a real corpus of **1,814 sessions / 90,626 tool calls**:

| | before | after |
|---|---|---|
| `tool_type` | `tool_use` 90,626 (single bucket) | `terminal_command` 48,008 · `tool_use` 24,617 · `mcp_tool` 18,001 |
| MCP servers attributed | 0 | 15 distinct servers |
| errors detected | **0** | 5,327 (5.9%), led by `Bash` (2,584) |

## Fix

Adds `ClaudeParser._classify_tool()`, mirroring what `opencode_parser._classify_tool()` already does for opencode — the Claude Code parser had simply not received the same treatment. Also:

- preserves `server_name` across the `tool_use` -> `tool_result` merge, which previously dropped it even when set;
- reads `is_error` from the tool result and maps it to `status="error"` plus the `error` field.

## Tests

Three new tests in `TestClaudeParser`: tool classification, server attribution surviving the result merge, and `is_error` producing `status="error"`.

`pytest tests` goes from 118 to 121 passing. The 5 `TestPlatformPathCoverage` failures in my environment are pre-existing and Windows-specific (confirmed unchanged with the patch stashed) — untouched by this PR.

### DIFF
--- a/Sensor/adr_sensor/parsers/claude_parser.py
+++ b/Sensor/adr_sensor/parsers/claude_parser.py
@@ -67,6 +67,24 @@ class ClaudeParser(BaseParser):
 
         return entries
 
+    def _classify_tool(self, tool_name: str) -> tuple:
+        """Classify a Claude Code tool call and attribute MCP tools to their server.
+
+        Claude Code namespaces MCP tools as ``mcp__<server>__<tool>``, so the server is
+        recoverable from the name alone. Without this, every call — including third-party
+        MCP servers — collapses into a single ``tool_use`` bucket with no server attribution.
+        """
+        if tool_name.startswith("mcp__"):
+            parts = tool_name.split("__")
+            if len(parts) >= 3 and parts[1]:
+                return "mcp_tool", parts[1]
+            return "mcp_tool", None
+
+        if tool_name in ("Bash", "PowerShell", "BashOutput", "KillShell"):
+            return "terminal_command", None
+
+        return "tool_use", None
+
     def _normalize_result_content(self, result_content: Any) -> str:
         """Normalize result content which can be a string or list of content items."""
         if isinstance(result_content, str):
@@ -186,7 +204,13 @@ class ClaudeParser(BaseParser):
 
                         if result_content and isinstance(result_content, str):
                             result_content = truncate_middle(result_content, max_length=1000, edge_chars=400)
-                        tool_results.append({"tool_use_id": tool_use_id, "result": result_content})
+                        tool_results.append(
+                            {
+                                "tool_use_id": tool_use_id,
+                                "result": result_content,
+                                "is_error": bool(item.get("is_error")),
+                            }
+                        )
 
             if tool_results:
                 extracted["tool_results"] = tool_results
@@ -245,14 +269,23 @@ class ClaudeParser(BaseParser):
                         for tool_result in tool_results:
                             tool_use_id = tool_result.get("tool_use_id")
                             result = tool_result.get("result")
+                            is_error = tool_result.get("is_error", False)
                             if tool_use_id in pending_tools:
                                 old_tool = pending_tools[tool_use_id]
+                                if is_error:
+                                    status = "error"
+                                elif result:
+                                    status = "success"
+                                else:
+                                    status = "unknown"
                                 updated_tool = ToolUsage(
                                     tool_name=old_tool.tool_name,
                                     tool_type=old_tool.tool_type,
+                                    server_name=old_tool.server_name,
                                     arguments=old_tool.arguments,
                                     result=result,
-                                    status="success" if result else "unknown",
+                                    status=status,
+                                    error=result if is_error else None,
                                 )
                                 for msg in entry.chat_history:
                                     if msg.role == "assistant":
@@ -274,9 +307,12 @@ class ClaudeParser(BaseParser):
                     tools = []
 
                     for tool_data in msg_data.get("tools", []):
+                        tool_name = tool_data.get("name", "unknown")
+                        tool_type, server_name = self._classify_tool(tool_name)
                         tool = ToolUsage(
-                            tool_name=tool_data.get("name", "unknown"),
-                            tool_type="tool_use",
+                            tool_name=tool_name,
+                            tool_type=tool_type,
+                            server_name=server_name,
                             arguments=tool_data.get("input", {}),
                             result=None,
                         )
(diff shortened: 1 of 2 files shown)


Sensor/adr_sensor/parsers/claude_parser.py
```diff
@@ -274,9 +307,12 @@ def _create_entry_from_extracted_session(
                     tools = []
 
                     for tool_data in msg_data.get("tools", []):
+                        tool_name = tool_data.get("name", "unknown")
```
pengyuzhang, line 310: **A `tool_use` block with `"name": null` now drops the entire session.**

`.get("name", "unknown")` only defaults when the key is *absent* — a present-but-null name yields `None`, and `_classify_tool(None)` raises `AttributeError` on `None.startswith("mcp__")`. That escapes to the broad `except Exception` in `_create_entry_from_extracted_session`, which prints and returns `None`, so the whole session — every user prompt and every other tool call in it — is silently discarded from telemetry. On `main` the same input parsed fine, since `tool_name=None` was simply stored.

Reproduced locally: the parser prints `[CLAUDE] Error creating entry for session s1` and returns `[]`.

```suggestion
                        tool_name = tool_data.get("name") or "unknown"
```

An `isinstance(tool_name, str)` guard at the top of `_classify_tool` would also cover the other call sites.

Sensor/adr_sensor/parsers/claude_parser.py
```diff
@@ -245,14 +269,23 @@ def _create_entry_from_extracted_session(
...
                             result = tool_result.get("result")
+                            is_error = tool_result.get("is_error", False)
                             if tool_use_id in pending_tools:
                                 old_tool = pending_tools[tool_use_id]
+                                if is_error:
+                                    status = "error"
+                                elif result:
+                                    status = "success"
+                                else:
+                                    status = "unknown"
```
pengyuzhang, line 280: **`status="unknown"` is provably wrong for empty non-error results.**

A `tool_result` block *is* the completion signal, and `is_error` is authoritative — anything with falsy `is_error` succeeded, whether or not its content normalizes to a non-empty string.

Measured against ~1,300 real tool calls in `~/.claude/projects`: 12 come out as `unknown`, all of them successful `ToolSearch` calls whose result content normalizes to `""`, each carrying an explicit `is_error: false` in the log. Collapsing the branch also matches the stated point of the PR:

```suggestion
                                status = "error" if is_error else "success"
```

Sensor/adr_sensor/parsers/claude_parser.py
```diff
@@ -245,14 +269,23 @@ def _create_entry_from_extracted_session(
...
                                     tool_type=old_tool.tool_type,
+                                    server_name=old_tool.server_name,
                                     arguments=old_tool.arguments,
                                     result=result,
-                                    status="success" if result else "unknown",
+                                    status=status,
+                                    error=result if is_error else None,
                                 )
                                 for msg in entry.chat_history:
                                     if msg.role == "assistant":
```
pengyuzhang, line 291: **The result merge two lines below matches by value, not identity — and this PR makes the misattribution worse.**

`ToolUsage` is `@dataclass(frozen=True)`, so two distinct pending calls with the same name/type/arguments compare equal while both still have `result=None`. `if t == old_tool` then hits whichever comes first, not the one this result belongs to.

Concrete failure: an assistant message issues two parallel `Bash` calls with the same command, and their `tool_result`s complete out of order (the erroring one first). The new `status="error"` and `error=…` land on the call that actually succeeded, and the success lands on the one that failed. Before this PR that only swapped result strings; now it mislabels *which call failed*. Value-identical repeated calls do occur in practice — 4 identical `ToolSearch` invocations in a ~1,300-call sample.

The sibling parser already fixes exactly this, at `claude_desktop_parser.py:428`:

```python
# Match on identity: two identical calls in different messages compare
# equal, and only the one this result belongs to should change.
if tool is not old_tool:
    continue
```

Separately, the inner `break` exits only the `for idx, t in enumerate(msg.tools)` loop, not the enclosing `for msg in entry.chat_history`, so a single result can be applied in more than one message.

pengyuzhang (commented):
Reviewed at commit 83bda7c. The premise checks out: `is_error` really does live on the `tool_result` content block (658 of 1296 blocks carry the key, 27 of them true in a ~1,300-call sample from `~/.claude/projects`), and the `mcp__<server>__<tool>` split handles names containing extra underscores, including the degenerate `mcp____x`. Test suite passes (76).

Three inline comments below — one of them a regression this PR introduces, so worth resolving before merge.

Two more findings that fall outside the diff, so they can't be anchored inline:

**`Sensor/adr_sensor/parsers/claude_desktop_parser.py:382` — the fix is half-applied across sources.** Claude Desktop agent mode is the most MCP-heavy source in the repo, yet its tool path still hardcodes `tool_type="tool_use"` with no `server_name`, and `_attach_tool_result` (~line 411) still does `status="success" if result else "unknown"` with no `is_error` handling and no `error` field. Every third-party MCP connector call from Claude Desktop stays in the unattributed bucket, and every failed call is still recorded as a success — the exact two gaps this PR closes for Claude Code.

**`Sensor/README.md:49` — undocumented behavior change.** The README documents the per-source MCP namespacing rule for opencode (`<server>_<tool>` → `mcp_tool` + `server_name`) but has no equivalent note for Claude Code, and the new `Bash`/`BashOutput`/`KillShell` → `terminal_command` reclassification is undocumented. Anyone reading the schema comment listing `tool_type` values has no way to know which ones Claude Code now emits.

Checked and cleared: `_classify_tool` handles `mcp__a_b___tool` correctly; `toolUseResult` is a plain string (`"Error: …"`) on every error case in the sampled logs, so the dict-only override at line 200 doesn't swallow error text; nothing in `Detection/` reads `tool_type` and `get_content_hash` has no non-test callers, so the `terminal_command` reclassification breaks no downstream consumer or event dedup.

PR #111, opened 2026-08-25
test(sensor): isolate redirected APPDATA test

weibit:

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**Related issue**: 

N/A

**What changed?**

Updated the redirected `%APPDATA%` Cursor parser test to use pytest's `monkeypatch` fixture and deterministic database candidates.

**Why?**

The test previously searched the normal macOS Cursor database before the temporary Windows database. On macOS systems with Cursor installed, the real database was selected and the test failed.

The revised test is isolated from applications and log files installed on the host machine.

**How did you test it?**

- `uv run pytest -q`
Result: `123 passed`
- `uv run ruff check adr_sensor tests`

**Potential risks**

Low. This changes test setup only and does not modify production parser behavior.

### DIFF
--- a/Sensor/tests/test_parsers.py
+++ b/Sensor/tests/test_parsers.py
@@ -1333,25 +1333,38 @@ class TestWindowsAppDataResolution:
 
     def test_redirected_appdata_end_to_end(self, tmp_path, monkeypatch):
         """A Cursor DB under a redirected %APPDATA% (outside the profile) is found."""
+        import importlib
+
+        from adr_sensor.parsers import cursor_parser as cursor_parser_module
+
         redirected = tmp_path / "fileserver/profiles$/alice/AppData/Roaming"
         db = redirected / "Cursor/User/globalStorage/state.vscdb"
         db.parent.mkdir(parents=True)
         db.touch()
 
-        monkeypatch.setenv("APPDATA", str(redirected))
+        try:
+            with monkeypatch.context() as isolated:
+                isolated.setenv("APPDATA", str(redirected))
+                # DB_PATHS is built at import time, so reload under the redirected env.
+                importlib.reload(cursor_parser_module)
 
-        windows_db = windows_appdata() / "Cursor/User/globalStorage/state.vscdb"
-        assert windows_db == db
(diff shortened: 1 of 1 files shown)


Sensor/tests/test_parsers.py
```diff
@@ -1331,27 +1331,27 @@ def test_localappdata_falls_back_to_profile_default(self, monkeypatch):
...
-            else:
-                os_mod.environ["APPDATA"] = original
-            importlib.reload(cursor_parser_module)
+        monkeypatch.setenv("APPDATA", str(redirected))
+
+        windows_db = windows_appdata() / "Cursor/User/globalStorage/state.vscdb"
+        assert windows_db == db
+
+        # Isolate this Windows-path test from real Cursor installations on the host.
+        monkeypatch.setattr(
```
pengyuzhang: This setup bypasses the production link between `%APPDATA%` and `CursorParser.DB_PATHS`: the test computes `windows_db` directly and then injects it into `DB_PATHS`. As a result, it still passes if `CursorParser` regresses to a hardcoded `~/AppData/Roaming` path—I verified that mutation with all 123 tests passing.

Could we keep the host isolation while testing the actual parser candidate, for example by reloading the parser after setting `APPDATA` and mocking path existence, or by retaining the reloaded `CursorParser.DB_PATHS` Windows entry and replacing only the macOS/Linux candidates?

weibit (commented):
Thanks for your review. You are right, the previous version bypassed the production APPDATA-to-DB_PATHS link. I updated the test to set APPDATA, reload cursor_parser, retain the Windows candidate generated by the reloaded CursorParser.DB_PATHS, and replace only the macOS/Linux candidates with nonexistent paths. The module is restored in a finally block. The full 123-test Sensor suite passed.

pengyuzhang (approved):
Reviewed the redirected APPDATA test isolation change. No blocking findings. Verified the exact PR head: 123 Sensor tests pass, and the changed test file passes Ruff.

weibit:
My apologies. I accidentally clicked “Close pull request.” Reopening it for merge.

Issue #50, opened 2026-08-19
Proposal: Structured triage and provenance-aware component analysis for ADR detection


shengqi-gensee:
## Summary

I would like to propose an extension to ADR Detection that gives Tier 1
**a structured way** to request deeper component investigation and gives Tier 2
**better provenance and decision semantics**.

The motivation is to improve recall for security-relevant component behavior
that is not evident from the transcript alone.

I have a working prototype and would like maintainers' feedback on the design
and preferred contribution boundaries before submitting implementation PRs.

## Motivation

ADR's fast/slow architecture is a good fit for this problem, but I encountered
three related limitations:

1. Tier 1 currently provides a mostly binary handoff. It cannot explicitly say
   that component behavior, object lineage, or source provenance is needed to
   resolve the risk.
2. Source inspection may cover only an entrypoint or invoked component, while
   relevant behavior can reside in imported implementation files or dormant
   registered components.
3. A single Tier-2 threat bit can conflate two different questions:
   - Is there evidence of a compromised component or malicious scenario?
   - Did the recorded trajectory produce an unauthorized effect?

These distinctions matter for cases where malicious component behavior is
present but was not exercised, an injected instruction was refused, or an
authorized/fixture workflow contains sensitive-looking operations without an
unauthorized effect.

### Structured Tier-1 handoff

Add an **optional** schema-constrained triage contract that reports:

- `BENIGN_EXIT` or `ESCALATE`;
- scenario/component risk;
- injection presence;
- unsafe plan, action, and observed effect;
- authorization status;
- evidence identifiers;
- whether deeper provenance is required, and why.

Malformed or internally inconsistent output would escalate rather than silently
exit. The existing stock text contract would remain the default.

### Bounded, label-free provenance

Allow Tier 2 to inspect source-bound component facts when requested:

- registered component identity and declared capabilities;
- source artifact digest;
- local imported implementation files;
- structural function/import information;
- explicit unavailable or incomplete coverage states.

The provider would return no malicious/benign verdict or benchmark label.
Paths and traversal would be constrained, and source content would be treated
as untrusted evidence rather than instructions.

### Separate component risk from observed effects

Optionally let Tier 2 report:

- `component_or_scenario_risk`
- `observed_unauthorized_effect`

Both would use `ABSENT`, `PRESENT`, or `UNRESOLVED`. Existing consumers could
continue receiving one compatibility decision derived from a configured axis.

The component-risk judgment would require concrete source, instruction,
[...]

PR #114, opened 2026-09-01
fix(detection): follow local imports in MCP source retrieval

shengqi-gensee:
**Related issue**:

Relates to https://github.com/uber/ADR/issues/50

**What changed?**

`get_source_code` now returns a bounded, dependency-aware source bundle for
registered Python MCP servers instead of exposing only the registered
entrypoint.

The implementation:

- statically follows local `import` and `from ... import ...` statements;
- includes package initializers and imported implementation modules;
- handles relative imports and namespace packages;
- follows Python package-versus-module and built-in/frozen/native-loader
  precedence;
- constrains resolved paths to the registered server directory;
- excludes symlinked source candidates;
- deduplicates cycles and preserves deterministic discovery order;
- never imports or executes discovered modules; and
- preserves the existing `source_code` entrypoint field and MCP input schema.

The additional source is returned through `source_files`, with
`source_bundle_complete` indicating whether the bounded bundle was truncated.

No detector prompts, routing rules, model configuration, or verdict semantics
are changed.

**How did you test it?**

Deterministic tests cover:

- local and relative imports;
- package initializers and `from package import submodule`;
- regular-package versus same-named-module precedence;
- namespace-package behavior;
- built-in and native-module precedence;
- cycles and unrelated-file exclusion;
- resolved-path containment and registry traversal attempts;
- symlink exclusion;
- confirmation that discovered modules are never executed; and
- preservation of the existing MCP schema and response shape.

Local results:

- Python 3.11 focused provider/baseline suite: 74 passed
- Python 3.12 focused provider/baseline suite: 74 passed
- Full Detection suite: 129 passed
- Registry smoke test: 115/115 registered sources retrieved

We also ran an isolated three-repeat ADR-Bench development comparison with
stock Tier 1 and Tier 2 behavior. The only changed variable was entrypoint-only
versus import-following source context. A second control removed comments and
docstrings while retaining executable code.

| Source control | Metric | Entrypoint only | Imports |
|---|---|---:|---:|
| Raw source | Mean recall | 77.8% | 82.5% |
| Raw source | Mean FPR | 13.5% | 12.9% |
| Comments/docstrings stripped | Mean recall | 77.8% | 84.1% |
| Comments/docstrings stripped | Mean FPR | 15.1% | 15.3% |

`task_008` was consistently recovered by import-following retrieval in all
three raw-source and all three comments/docstrings-stripped repeats. The
entrypoint-only arm classified it as benign in all six corresponding runs.

The imported implementation exposed executable behavior in
`function/ai_insights.py` and `function/system_analyzer.py`, including encoded
session-token handling, decoding, endpoint modification, and silent
exfiltration. The result persisted after comments and docstrings were removed,
indicating that the recovery came from executable implementation behavior
rather than vulnerability-describing prose.

### DIFF
--- a/Detection/context_providers/source_code_analyzer_server.py
+++ b/Detection/context_providers/source_code_analyzer_server.py
@@ -5,11 +5,19 @@ ADR Source Code Context Provider
 Provides MCP server source code for reasoning-based threat analysis.
 No pre-analysis or cheating metadata - just clean source code and basic info.
 """
-import yaml
+import ast
 import logging
+import os
+from importlib.machinery import (
+    BYTECODE_SUFFIXES,
+    EXTENSION_SUFFIXES,
+    BuiltinImporter,
+    FrozenImporter,
+)
 from pathlib import Path
-from typing import Dict, List, Any
+from typing import Any, Dict, List, Optional, Tuple
 
+import yaml
 from mcp.server.fastmcp import FastMCP
 
 # Configure logging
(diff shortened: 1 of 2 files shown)


barisozbas (changes requested):
Hi, Please see the below findings. 

 - High — source_bundle_complete
(https://github.com/uber/ADR/blob/241d4c57040b098a262aef7a2ae4404271895efa/Detection/context_providers/source_code_analyzer_server.py#L325)
    incorrectly remains true when the 24-file limit omits queued dependencies. I reproduced 25 files returning only 24
    while reporting complete. Exact size-boundary exhaustion has the same problem.

  - Medium — The collector
(https://github.com/uber/ADR/blob/241d4c57040b098a262aef7a2ae4404271895efa/Detection/context_providers/source_code_analyzer_server.py#L255)
    reads entire files before applying the limit and counts characters, not bytes. A 100,002-character source produced
    400,002 UTF-8 bytes despite the 240,000 budget.

  - Medium — A valid non-UTF-8 Python dependency raises UnicodeDecodeError, causing the entire server response—including ts entrypoint—to become status: error.

shengqi-gensee:
Thanks for the review—all three findings were valid and are now fixed:

- Bundle completeness accounts for dependencies omitted by file or byte limits.
- Reads are bounded and budgeted in bytes, including failed decoding attempts.
- Python encoding declarations are supported; invalid codecs or dependency decoding failures preserve the entrypoint and mark the bundle incomplete.

Added regression coverage for these cases and truncation boundaries. All 145 tests pass. Ready for re-review.

barisozbas (approved):
Comments resolved. LGTM

PR #119, opened 2026-09-09
fix(sensor): honor CLI display limit

### LINKED ISSUE #118: Sensor CLI --limit option has no effect
## Description

The Sensor CLI advertises `--limit` as the "Number of entries to display", but the parsed value is never consumed. Changing the option therefore has no effect on output.

## Steps to reproduce

Run the Sensor against the same available logs with different limits:

```sh
adr-sensor --limit 0 --no-save
adr-sensor --limit 10 --no-save
```

## Actual behavior

Both commands print the same aggregate ingestion table. No individual entry summaries are displayed, and `args.limit` is not referenced after argument parsing.

## Expected behavior

The CLI should display at most the requested number of entry summaries. A limit of zero should suppress the entry preview, and negative values should be rejected.

## Environment

Observed on current `main` (`df05577`). The defect is platform-independent because it is in CLI argument wiring.

## Proposed fix

Pass the parsed limit into `AgentObserver.display_summary()`, render a newest-first bounded preview using the existing `AgentEvent.get_summary()`, and add CLI and observer regression tests.

sanjayy0612:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

**Related issue**: Closes #118

**What changed?**

- Pass the parsed `--limit` value into `AgentObserver.display_summary()`.
- Print a newest-first, bounded preview using the existing `AgentEvent.get_summary()` representation.
- Treat `--limit 0` as disabling the preview and reject negative values during argument parsing.
- Add CLI wiring, validation, ordering, and limit regression tests.

**Why?**

The CLI advertised `--limit` as controlling the number of displayed entries, but the value was never consumed. Every value therefore produced the same aggregate-only output.

**How did you test it?**

- `.venv/bin/pytest -q` — 128 passed
- `.venv/bin/ruff check adr_sensor/`
- Ruff checks on the changed test files
- `git diff --check`

**Potential risks**

Low. The default CLI output now includes the two entry summaries the existing default limit promised. API callers can pass `limit=0` to retain aggregate-only output. Entry ordering is newest first and normalizes naive timestamps as UTC through the existing timestamp utility.

### DIFF
--- a/Sensor/adr_sensor/cli.py
+++ b/Sensor/adr_sensor/cli.py
@@ -31,6 +31,14 @@ def get_version():
     return __version__
 
 
+def _non_negative_int(value: str) -> int:
+    """Parse a non-negative integer for bounded display options."""
+    parsed = int(value)
+    if parsed < 0:
+        raise argparse.ArgumentTypeError("must be zero or greater")
+    return parsed
+
+
 def main():
     """Main entry point for the ADR Sensor CLI."""
     parser = argparse.ArgumentParser(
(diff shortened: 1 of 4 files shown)


barisozbas (approved):
lgtm

PR #123, opened 2026-09-13
feat(sensor): add DeepSeek Harness parser

DvJiang:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

**Related issue**: N/A

**What changed?**

- Add a DeepSeek Harness parser for current v3 JSONL and Zstandard-compressed session logs.
- Normalize messages, tool calls and results, PTC sub-dispatches, approvals, provider/model context, and token usage into the Sensor schema.
- Preserve structured tool-result content and metadata, including compaction replacement provenance, while keeping the original execution result in normalized tool telemetry.
- Register the `dsh` source with `AgentObserver` and the CLI, including content-aware incremental exports for resumed sessions.
- Add parser documentation, a three-platform CI workflow, runtime `zstandard` dependency, and focused tests.

**Why?**

DeepSeek Harness stores local agent sessions in a structured event stream that ADR Sensor did not previously ingest. This adds read-only support using the same source-specific parser pattern as the existing agent integrations.

**How did you test it?**

- `pytest tests/ -q` on Python 3.11: 245 passed.
- `pytest tests/test_dsh_parser.py -q` on Python 3.9: 14 passed.
- Ruff checks and formatting checks pass for the changed parser and tests.
- `uv lock --check` passes.
- Source distribution and wheel builds complete successfully.

**Potential risks**

DeepSeek Harness may evolve its on-disk event schema. To avoid silently misinterpreting incompatible data, the parser only accepts v3 as the highest session generation and skips legacy or future generations. It reads session files without modifying them. For a torn trailing Zstandard frame, it follows the upstream recovery contract by retaining only complete newline-terminated JSONL records and discarding the incomplete tail.


barisozbas (changes requested):
This adds useful DeepSeek Harness support, but I recommend addressing these issues before merging:

  ### 1. Preserve tool-result metadata and replacement history

  [Parser line 257](https://github.com/uber/ADR/blob/8e0b61bdd01ae8c70b39f8906927587d9656b6da/Sensor/adr_sensor/parsers/
  dsh_parser.py#L257)

  Tool results are reduced to text/status/error, dropping attachment references and `meta`, which can contain contextual
  filesystem diffs.

  DSH compaction can also append a replacement result with the same call ID. The parser overwrites the original
  execution result without retaining the replacement relationship.

  Please retain structured result metadata and distinguish execution results from later model-context replacements. This
  is separate from the text truncation already used by other Sensor parsers.

  ### 2. Match upstream recovery for incomplete Zstandard tails

  [Parser line 434](https://github.com/uber/ADR/blob/8e0b61bdd01ae8c70b39f8906927587d9656b6da/Sensor/adr_sensor/parsers/
  dsh_parser.py#L434)

  The parser discards all decoded content from an incomplete final frame. However, DSH’s persistence contract recovers
  complete newline-terminated JSONL records from that frame.

  A frame missing only its final checksum byte can contain a fully decoded event that ADR currently drops.

  Please recover complete records from the incomplete final frame and update the test that currently expects them to be
  discarded.

  ### 3. Preserve distinct source session IDs

  [Parser line 187](https://github.com/uber/ADR/blob/8e0b61bdd01ae8c70b39f8906927587d9656b6da/Sensor/adr_sensor/parsers/
  dsh_parser.py#L187)

  Stripping the optional `session-` prefix maps both `session-review` and `review` to `dsh_review`. These are valid
  distinct custom/imported IDs, but the parser deduplicates them into one session.

  Please namespace the complete source ID without stripping its prefix and add a regression test covering this
  collision.

DvJiang:
Thanks for the detailed review. I addressed all three points in 847c0b4:

1. Tool-result content, attachments, `meta`, errors, `surfaceOp`, and `sourceEventSeqs` are retained in `session_context.tool_result_metadata`. Compaction replacements no longer overwrite the normalized original execution result.
2. Torn trailing Zstandard frames now recover complete newline-terminated JSONL records, including the missing-checksum-byte case, while discarding only the incomplete record tail.
3. DSH session IDs now namespace the complete source ID, with a regression test covering `session-review` and `review` as distinct sessions.

The full Sensor suite passes on Python 3.11 (245 tests), the focused DSH suite passes on Python 3.9 (14 tests), and Ruff, lockfile, and package-build checks pass locally.

barisozbas (approved):
open findings resolved. LGTM

